PR #43, opened 2020-07-10
test: added retryUntilExists in claimValue page (CMC-NA)

swalker125:

N/A

### Change description ###

The e2e fails often on the claim value page, due to the callback not returning quickly enough. Placed retry until exist in claim value page

```
[ ] Yes
[ ] No
```


sabahirfan (dismissed):
LGTM 👍

PR #49, opened 2020-07-14
fix: make lambok builder play nicely with Jackon mapper

ckalista:
### Change description ###

Make Lambok builder work with Jackson mapper. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/ucmc/model/ClaimValue.java
+++ b/src/main/java/uk/gov/hmcts/reform/ucmc/model/ClaimValue.java
@@ -2,14 +2,16 @@ package uk.gov.hmcts.reform.ucmc.model;
 
 import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 import lombok.RequiredArgsConstructor;
 
 @Data
 @Builder
 @RequiredArgsConstructor
+@NoArgsConstructor(force = true)
 public class ClaimValue {
-    private final Integer lowerValue;
-    private final Integer higherValue;
+    private  final Integer lowerValue;
+    private  final Integer higherValue;
 
     public boolean hasLargerLowerValue() {
         if (lowerValue == null || higherValue == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/ucmc/model/ClaimValue.java
```diff
@@ -2,11 +2,13 @@
 
 import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 import lombok.RequiredArgsConstructor;
 
 @Data
 @Builder
 @RequiredArgsConstructor
+@NoArgsConstructor(force = true)
```
sabahirfan, line 11: Have you tried constructor with @JasonCreator annotation on top of that? 
```suggestion
@JsonCreator
    public ClaimValue(Integer lowerValue, Integer higherValue) {
        this.lowerValue = lowerValue;
        this.higherValue = higherValue;
    }
```

sabahirfan: With lombok it will be: `@RequiredArgsConstructor(onConstructor = @__(@JsonCreator))`

ckalista: Easy either way - just thought the whole point was to avoid having to explicitly declare the constructor which can get quite verbose the more fields you have in your builder...?

ckalista: Just tried suggestion above but unfortunately: 

`java.lang.IllegalArgumentException: Invalid type definition for type `uk.gov.hmcts.reform.ucmc.model.ClaimValue`: Argument #0 has no property name, is not Injectable: can not use as Creator [constructor for uk.gov.hmcts.reform.ucmc.model.ClaimValue, annotations: {interface com.fasterxml.jackson.annotation.JsonCreator=@com.fasterxml.jackson.annotation.JsonCreator(mode=DEFAULT)}]
 at [Source: UNKNOWN; line: -1, column: -1]`

ckalista:
image

Setting the build tool in IntelliJ to using Gradle instead of default gets around the problem. Not quite sure what is going on here. 

Closing this for now.

PR #50, opened 2020-07-15
feat: added case stayed event and logic (CMC-336)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-336

### Change description ###

New MOVE_TO_STAYED event
New CoreCaseDataService with triggerEvent and searchCase methods
New CaseStayedFinder which will be triggered from Quartz

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/ucmc/service/CoreCaseDataService.java
```diff
@@ -0,0 +1,66 @@
...
+
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class CoreCaseDataService {
+    private final IdamClient idamClient;
+    private final CoreCaseDataApi coreCaseDataApi;
+    private final SystemUpdateUserConfiguration userConfig;
+    private final AuthTokenGenerator authTokenGenerator;
+
+    public static final String CASE_TYPE = "UNSPECIFIED_CLAIMS";
```
ckalista: I know very unlikely to ever change but still shall we keep this in a config?

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchService.java
```diff
@@ -0,0 +1,29 @@
...
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class CaseSearchService {
+    private final CoreCaseDataService coreCaseDataService;
+
+    public List<CaseDetails> getCasesOver112Days() {
+        return coreCaseDataService.searchCases(dateQuery());
+    }
+
+    private String dateQuery() {
+        final String dateProperty = "data.claimIssuedDate";
```
ckalista: `claimIssuedDate` has not yet been populated...Don't think we have a user story to cover that!

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseStayedFinder.java
```diff
@@ -0,0 +1,38 @@
...
+        log.info("Job {} started", jobName);
+
+        List<CaseDetails> cases = caseSearchService.getCasesOver112Days();
+
+        if (cases.isEmpty()) {
+            log.info("Job '{}' did not find any cases", jobName);
+        } else {
+            log.info("Job '{}' found {} case(s)", jobName, cases.size());
+
+            cases.forEach(caseDetails -> applicationEventPublisher.publishEvent(
```
ckalista: 👍

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchService.java
```diff
@@ -0,0 +1,29 @@
...
+        final Map<String, Object> dayRange = of("gt", "now+112d");
+
+        return new JSONObject(of("query", of("range", of(dateProperty, dayRange)))).toString();
```
ckalista: Can we confirm what is the default in terms of pagination? We may want to have a think about how we would handle a huge backlog of matching cases.

ckalista: Elastic Search allows map-reduce: https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=843514186#CCDElasticSearchandnewsearchAPIDesignLLD-Returningonlyasubsetofthecaseinformation

We are only interested in case ids, worth exploring how hard would it be to optimise here?

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseStayedFinder.java
ckalista: I like the approach - could you just double check the application publisher is synchronous. We want to to be absolutely sure that each `MOVE_TO_STAYED` event sent to CCD is blocking before sending the next one 👍

ckalista (dismissed):
Looking really good - just did a partial review with a few comments just to tease our brains :-)

src/test/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchServiceTest.java
```diff
@@ -0,0 +1,53 @@
...
+    private ArgumentCaptor<String> queryCaptor;
+
+    @Mock
+    private CoreCaseDataService coreCaseDataService;
+
+    @InjectMocks
+    private CaseSearchService searchService;
+
+    @Test
+    void shouldSearchCasesByDateProperty() throws JSONException {
```
ckalista: We filed a story (CMC-502) to introduce naming conventions of test methods - let's try for all new work to stick to that already:
https://www.petrikainulainen.net/programming/testing/writing-clean-tests-naming-matters/

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchService.java
swalker125: I started adding the definition for claimIssuedDate, but you're right all of that could actually be lifted out of this and raised as a separate PR with a callback to populate

src/main/java/uk/gov/hmcts/reform/ucmc/service/CoreCaseDataService.java
swalker125: sure can move to config

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseStayedFinder.java
swalker125: yep standard EventListener is syncronous by default

src/main/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchService.java
swalker125: this is a really good question, by default ES only returns 10 results. We can overwrite this using size but how big should that size be? I have no idea 🤔 I don't think many cases will ever not be served, might be worth getting some figures?

swalker125: as for returning case data, you're right I do not need to do that. I can use data from the startEventResponse when I call the trigger event method in corecasedata service. I will look into returning just case id

src/test/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchServiceTest.java
swalker125: it's still not 100% clear to me how tests should be formatted

swalker125:
Query 

```
{
    "query":{
        "range":{
            "data.claimIssuedDate": {
                "lt": "now-112d"
            }
        }
    },
    "_source": ["reference"]
} 
```

Results

```
{
    "took": 3,
    "timed_out": false,
    "_shards": {
        "total": 2,
        "successful": 2,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": 1,
        "max_score": 1.0,
        "hits": [
            {
                "_index": "unspecified_claims_cases-000001",
                "_type": "_doc",
                "_id": "1",
                "_score": 1.0,
                "_source": {
                    "reference": 1594913107398687
                }
            },
        ]
    }
}
```

src/main/java/uk/gov/hmcts/reform/unspec/handler/MoveCaseToStayedEventHandler.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Slf4j
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class MoveCaseToStayedEventHandler {
+    private final CoreCaseDataService coreCaseDataService;
+
+    @EventListener
+    public void moveCaseToStayed(MoveCaseToStayedEvent event) {
+        coreCaseDataService.triggerEvent(event.getCaseId(), "MOVE_TO_STAYED");
```
pd-dudkiewicz: should we store event names in enum?

src/main/java/uk/gov/hmcts/reform/unspec/service/CaseSearchService.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public List<CaseDetails> getCasesToBeStayed() {
+        return coreCaseDataService.searchCases(dateQuery()).getCases();
+    }
+
+    private String dateQuery() {
+        final String dateProperty = "data.claimIssuedDate";
+        final Map<String, Object> dayRange = of("lt", "now-112d");
+
+        return new JSONObject(of("query", of("range", of(dateProperty, dayRange)), "_source", List.of("reference")))
```
pd-dudkiewicz: How about formatting it like a json so it's easier to read?
```
return new JSONObject(
    of("query",
       of("range",
          of("data.claimIssuedDate",
             of("lt", "now-112d")
          )
       ),
    "_source", List.of(
        "reference"
    )
)).toString();
```

src/main/java/uk/gov/hmcts/reform/unspec/service/CaseStayedFinder.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public void execute(JobExecutionContext context) {
+        final String jobName = context.getJobDetail().getKey().getName();
+        log.info("Job {} started", jobName);
+
+        List<CaseDetails> cases = caseSearchService.getCasesToBeStayed();
+
+        if (cases.isEmpty()) {
+            log.info("Job '{}' did not find any cases", jobName);
+        } else {
```
pd-dudkiewicz: Do we need that part? It could simply log "found 0 case(s)".

src/main/java/uk/gov/hmcts/reform/unspec/service/CoreCaseDataService.java
```diff
@@ -0,0 +1,65 @@
...
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class CoreCaseDataService {
+    private final IdamClient idamClient;
+    private final CoreCaseDataApi coreCaseDataApi;
+    private final SystemUpdateUserConfiguration userConfig;
+    private final AuthTokenGenerator authTokenGenerator;
+
+    public static final String CASE_TYPE = "UNSPECIFIED_CLAIMS";
+    public static final String JURISDICTION = "CIVIL";
```
pd-dudkiewicz: might be good idea to have both on some enum/const

src/main/resources/application.yaml
```diff
@@ -36,13 +36,54 @@ spring:
 oidc:
   issuer: ${OIDC_ISSUER:http://fr-am:8080/openam/oauth2/hmcts}
 
+idam:
+  s2s-auth:
+    microservice: unspec_service
+
+unspecified:
+  response-pack-url: https://formfinder.hmctsformfinder.justice.gov.uk/n9-eng.pdf
+  system_update:
```
pd-dudkiewicz: Should we decide to stick to either dashes od underscores for config properties names (I think the general rule is to use dashes)? Wherever we have control over it obviously, because it differs in dependencies we use (e.g. for ccd client it needs to be core_case_data).

src/test/java/uk/gov/hmcts/reform/ucmc/service/CaseSearchServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+        String property = "data.claimIssuedDate";
+
```
pd-dudkiewicz: Can we get rid of empty lines, so we only use them to separate arrange, act, assert steps?

src/test/java/uk/gov/hmcts/reform/ucmc/service/CaseStayedFinderTest.java
```diff
@@ -0,0 +1,76 @@
...
+        verify(applicationEventPublisher).publishEvent(new MoveCaseToStayedEvent(caseId));
+    }
+
+    @Test
+    void shouldNotEmitMoveCaseToStayedEventWhenNoCasesFound() {
+        when(caseSearchService.getCasesToBeStayed()).thenReturn(List.of());
+
+        caseStayedFinder.execute(jobExecutionContext);
+
+        verifyZeroInteractions(applicationEventPublisher);
```
pd-dudkiewicz: the method is deprecated, `verifyNoInteractions` should be used instead

src/test/java/uk/gov/hmcts/reform/ucmc/service/CaseStayedFinderTest.java
```diff
@@ -0,0 +1,76 @@
+package uk.gov.hmcts.reform.unspec.service;
```
pd-dudkiewicz: Package name is corrrect, but you've got your tests in `ucmc` directory.

src/main/resources/application.yaml
swalker125: will change to -

src/main/java/uk/gov/hmcts/reform/unspec/service/CaseStayedFinder.java
```diff
@@ -24,11 +24,9 @@ public void execute(JobExecutionContext context) {
 
         List<CaseDetails> cases = caseSearchService.getCasesToBeStayed();
 
-        if (cases.isEmpty()) {
-            log.info("Job '{}' did not find any cases", jobName);
-        } else {
-            log.info("Job '{}' found {} case(s)", jobName, cases.size());
+        log.info("Job '{}' found {} case(s)", jobName, cases.size());
 
+        if (!cases.isEmpty()) {
```
pd-dudkiewicz: this is probably not needed, `forEach` just won't do anything for empty list

pd-dudkiewicz (dismissed):
one more minor comment, other than that looks good to me

charts/unspec-service/Chart.yaml
```diff
@@ -2,6 +2,6 @@ apiVersion: v2
 description: A Helm chart for unspec-service App
 name: unspec-service
 home: https://github.com/hmcts/unspec-service
-version: 1.0.0
+version: 1.0.1
```
pd-dudkiewicz: Should it be 1.1.0? We're adding new stuff so it's not bugfix.

pd-dudkiewicz (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/service/CaseSearchService.java
```diff
@@ -0,0 +1,34 @@
...
+    public List<CaseDetails> getCasesToBeStayed() {
+        return coreCaseDataService.searchCases(dateQuery()).getCases();
+    }
+
+    private String dateQuery() {
+        return new JSONObject(
+            of("query",
+                of("range",
+                    of("data.claimIssuedDate",
+                        of("lt", "now-112d")
```
douglasrice, line 28: now-112d: as we are searching for records which are 4 months old would now-4M be better?

src/main/java/uk/gov/hmcts/reform/unspec/service/CaseSearchService.java
```diff
@@ -0,0 +1,34 @@
...
+                    )
```
douglasrice, line 29: additional filter required for the additional AC added by Chris:
And the claim is still in CREATED state

douglasrice (changes requested):

douglasrice (approved):

pd-dudkiewicz (approved):

PR #76, opened 2020-07-23
impr: added pagination to es queries (CMC-336)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-336

### Change description ###

ES only returns 10 cases by default. Rather than add size param we will keep querying 10 results at a time.

Moved search services into search directory under services

This PR also adds state check to existing case stayed filter

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/unspec/service/search/ElasticSearchService.java
```diff
@@ -0,0 +1,33 @@
...
+    private final CoreCaseDataService coreCaseDataService;
+
+    private static final int startIndex = 0;
+    private static final int esDefaultSearchLimit = 10;
+
+    public List<CaseDetails> getCases() {
+        SearchResult searchResult = coreCaseDataService.searchCases(query(startIndex));
+        List<CaseDetails> caseDetails = new ArrayList<>(searchResult.getCases());
+
+        while (searchResult.getTotal() > caseDetails.size()) {
```
ckalista: That's dangerous. If for whatever reason subsequent ES queries don't return any result (most likely won't happen but still...) we are at risk of looping endlessly which could cause a DOS on ES but also break our own service 💥 :)

Safer approach is a good old for loop, working out how many pages we have got 👍

src/main/java/uk/gov/hmcts/reform/unspec/service/search/CaseStayedSearchService.java
```diff
@@ -0,0 +1,32 @@
...
+public class CaseStayedSearchService extends ElasticSearchService {
+
+    public CaseStayedSearchService(CoreCaseDataService coreCaseDataService) {
+        super(coreCaseDataService);
+    }
+
+    //TODO: update case data to store claimIssuedDate + 4 months. ES query can then be done on this value using now
+    public String query(int startIndex) {
+        return new JSONObject(
+            of("query",
```
ckalista: We could probably try to build a simple reusable mechanism to handle paging completely transparently. There is a full query DSL for ES that could be used to build the basic query that can then be augmented using composition. eg I could imagine something like this: `fetchAll(dslQuery)`

https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/java-query-dsl.html

ckalista: Actually in CMC we used that DSL: https://github.com/hmcts/cmc-claim-store/blob/d76f3e9bb3f0228dcde1cf871b17a9cece7e61f4/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDElasticSearchRepository.java#L44

ckalista (changes requested):
We probably will have to revisit the potential risk of endless loop. Other than that the other comment is more a suggestion for improvement to consider introducing some sort of generic pagination handling mechanism since we will need it again and again...can do it later when we encounter the need, up to you 👍

src/main/java/uk/gov/hmcts/reform/unspec/service/search/ElasticSearchService.java
swalker125: haha I was up late thinking exactly this last night 😅

src/main/java/uk/gov/hmcts/reform/unspec/service/search/CaseStayedSearchService.java
swalker125: I can't stand this horrible map of of of of of so yes, wasn't aware of this. Shall implement

douglasrice (approved):

dharmendrak (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/service/search/ElasticSearchService.java
```diff
@@ -0,0 +1,34 @@
...
+    private final CoreCaseDataService coreCaseDataService;
+
+    private static final int startIndex = 0;
+    private static final int esDefaultSearchLimit = 10;
+
+    public List<CaseDetails> getCases() {
+        SearchResult searchResult = coreCaseDataService.searchCases(query(startIndex));
+        List<CaseDetails> caseDetails = new ArrayList<>(searchResult.getCases());
+
+        for (int i = esDefaultSearchLimit; caseDetails.size() < searchResult.getTotal(); i = i + esDefaultSearchLimit) {
```
ckalista: Still potentially have a risk here in case ES query doesn't return any record subsequently. Discussing this on slack... 👍

douglasrice (dismissed):

ckalista (changes requested):

ckalista (approved):

douglasrice (approved):

PR #81, opened 2020-07-27
fix: set the same OIDC_ISSUER for aat and preview

pd-dudkiewicz:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -1,6 +1,12 @@
 server:
   port: 4000
 
+logging:
+  level:
+    org:
+      springframework:
+        security: DEBUG
+
 management:
   health:
     probes:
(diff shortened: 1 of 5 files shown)


src/main/resources/application.yaml
```diff
@@ -1,6 +1,12 @@
 server:
   port: 4000
 
+logging:
+  level:
+    org:
+      springframework:
+        security: DEBUG
```
sabahirfan, line 8: I suppose that is temporary and will be removed later on.

sabahirfan (dismissed):
LGTM

src/main/resources/application.yaml
pd-dudkiewicz: yes, it will

dharmendrak (dismissed):

ckalista (approved):

sabahirfan (approved):

PR #105, opened 2020-08-04
feat: claim issue date, deemed date of service, response deadline (CMC-510)

sabahirfan:
[CMC-510](url)

### Change description ###
This PR implements the calculation of issue date, deemed date of service, response deadline and confirmation of service dates.

```
[ ] Yes
[x] No
```


ccd-definition/CaseField.json
```diff
@@ -89,6 +89,22 @@
     "FieldType": "Date",
     "SecurityClassification": "Public"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "ID": "claimSubmittedDateTime",
```
swalker125, line 95: why is this field needed as well as claimIssueDate? Are they not the same?

swalker125 (commented):
will look through the rest later today, just a question around what claimSubmittedDateTime is for

ccd-definition/CaseField.json
sabahirfan: Nope these are two different dates. `claimSubmittedDateTime` is the timestamp when claim is submitted and required on the sealed claim pdf, while `claimIssueDate ` can be next working day and after 4pm etc.

swalker125: are you sure? claimIssuedDate is just when I submit the form right? I don't see why that needs to be next working day etc. In that case they are the same. If not the same, why do we need to store this at all? Is it not throw away info?

dharmendrak (approved):

ccd-definition/CaseField.json
sabahirfan: yes, I am sure. look at the story for requirements for this.
```
IssueDate =
now
+ 1 day if >= 16:00 PM
+ 1 day till next working day
```

swalker125: then claimIssueDate can probably be removed and we will just use claimIssueDateAndTime? I don't see why we need both fields

sabahirfan: As mentioned earlier both dates are needed on sealed claim PDF.

karthikkannan-lemuria (approved):

PR #146, opened 2020-08-27
test: fast fail e2e tests on failure (CMC-NA)

swalker125:

N/A

### Change description ###

Added mocha fail:true config to codecept.js.conf file

```
[ ] Yes
[x] No
```


dharmendrak (approved):

sabahirfan (commented):
I am not sure about the requirement here. 
In my opinion it would be nice to see how many tests actually failed for a change. Rather than finding one by one. @ckalista please advise.

pd-dudkiewicz (approved):

swalker125:
@sabahirfan this is for e2es. When one test fails, all subsequent tests fail too as they're built on top of each other. I then need to wait extra time for these all to fail. So seeing that 5 tests failed is usually misleading. What do you think?

sabahirfan (approved):

PR #203, opened 2020-09-16
fix: case name in claim pdf (CMC-740)

sabahirfan:
[CMC-740](https://tools.hmcts.net/jira/browse/CMC-740)

### Change description ###
This PR fixes the case name in sealed claim pdf.

- [ ] Yes
- [x] No


dharmendrak (approved):

pd-dudkiewicz (approved):

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandler.java
```diff
@@ -118,11 +118,15 @@ private CallbackResponse issueClaim(CallbackParams callbackParams) {
...
+        Party respondent = updatePartyWithPartyName(caseDetails.getData().get(RESPONDENT));
+        Party applicant = updatePartyWithPartyName(caseDetails.getData().get(CLAIMANT));
         CaseDocument sealedClaim = sealedClaimFormGenerator.generate(
             caseData.toBuilder()
                 .claimIssuedDate(issueDate)
                 .legacyCaseReference(referenceNumber)
                 .claimSubmittedDateTime(submittedAt)
+                .applicant1(applicant)
+                .respondent1(respondent)
                 .build(),
```
swalker125, line 130: hmm I won't stop as it fixes, but I don't like that we pass in this caseData.toBuilder() stuff. In my mind the generate method should map case details -> case data and then it has the values available. The generate method could then take in case details and a date (in this case submittedAt - which again I think makes sense as all documents will need a generated date). Then this code becomes a lot simpler

swalker125: I think the transform of respondent and claimant could happen in the case data convertor too, then it's avaiable in the generate method and can be added to data with data.put(RESPONDENT, case.getRespondent);

swalker125: made a tech debt ticket here https://tools.hmcts.net/jira/browse/CMC-750

douglasrice (approved):

PR #218, opened 2020-09-22
feat: generic service notification (CMC-636)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/CMC-639

### Change description ###

Added generic notification to inform solicitors about response to case.

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseEvent/systemUpdate.json
```diff
@@ -12,5 +12,47 @@
     "CaseEventID": "NOTIFY_DEFENDANT_SOLICITOR_FOR_CLAIM_ISSUE",
     "UserRole": "caseworker-civil-systemupdate",
     "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "CaseEventID": "NOTIFY_CLAIMANT_SOLICITOR_FOR_REQUEST_FOR_EXTENSION",
```
swalker125: Might be worth looping in Alison for these event names, I don't know if they will need to be specific to each solicitor for multiparty

ccd-definition/CaseEvent.json
```diff
@@ -156,7 +156,7 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "UNSPECIFIED_CLAIMS",
     "ID": "NOTIFY_DEFENDANT_SOLICITOR_FOR_CLAIM_ISSUE",
-    "Name": "Notify defendant solicitor",
+    "Name": "Notify claim issue",
```
swalker125, line 164: I'm not sure we can be as generic as this, I think we will need a different event for each email

swalker125 (commented):
made a couple of comments on the 639 pr that are relevant to this, won't comment again so will wait for Dharmendra to get back to me

ccd-definition/CaseEvent.json
pd-dudkiewicz: Might be a good idea to have them separated in ccd. Do you also think that Camunda should be sending list of ccd events to trigger in single service task, or would it be better to trigger each one separately?

swalker125: I believe we've said it will be indivudal tasks so we can accurately know what has and hasn't happened on a case. @ckalista will be able to inform you better here I am sure :)

ckalista: Yeah it needs to be a seperate Service Task per CCD Event or else how would we recover if say event II in the list failed in CCD...?

When something goes rogue we need to be able to recover from the situation safely.
there will be default retries - if you handle multiple events per Service Task if once of the event fails, the service task would retry all of the events…

dharmendrak: we agreed in past that we will have same External task listener used with Case_Event name hard coded in parameters of specific Camunda Task

dharmendrak (dismissed):
LGTM, however we can make the notification handler bit in more generic handler ways, better to do later

ccd-definition/AuthorisationCaseEvent/systemUpdate.json
AlisonRevittHM: @pd-dudkiewicz as discussed, for this event it can be NOTIFY_APPLICANT_SOLICITOR1_FOR_REQUEST_FOR_EXTENSION. If you need to underscore before the integer then go for it. We need to review all the other events to future proofing.

ccd-definition/AuthorisationCaseEvent/systemUpdate.json
```diff
@@ -12,5 +12,61 @@
...
+    "UserRole": "caseworker-civil-systemupdate",
+    "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "CaseEventID": "NOTIFY_DEFENDANT_SOLICITOR_FOR_EXTENSION_RESPONSE",
```
AlisonRevittHM: NOTIFY_RESPONDENT_SOLICITOR1_FOR_EXTENSION_RESPONSE

ccd-definition/AuthorisationCaseEvent/systemUpdate.json
```diff
@@ -12,5 +12,61 @@
...
+    "UserRole": "caseworker-civil-systemupdate",
+    "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "CaseEventID": "NOTIFY_CLAIMANT_SOLICITOR_FOR_SERVICE_ACKNOWLEDGEMENT",
```
AlisonRevittHM: NOTIFY_APPLICANT_SOLICITOR1_FOR_SERVICE_ACKNOWLEDGEMENT

ccd-definition/AuthorisationCaseEvent/systemUpdate.json
```diff
@@ -12,5 +12,61 @@
...
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "CaseEventID": "NOTIFY_DEFENDANT_SOLICITOR_FOR_CASE_TRANSFERRED_TO_LOCAL_COURT",
+    "UserRole": "caseworker-civil-systemupdate",
+    "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "CaseEventID": "NOTIFY_CLAIMANT_SOLICITOR_FOR_CASE_TRANSFERRED_TO_LOCAL_COURT",
```
AlisonRevittHM: Can you use APPLICANT / RESPONDENT and also add the integers for the appropriate events based on the discussion today?

AlisonRevittHM (changes requested):
data review

ccd-definition/AuthorisationCaseEvent/systemUpdate.json
pd-dudkiewicz: all done 👍

dharmendrak (dismissed):

PR #235, opened 2020-09-25
feat: event management (CMC-545)

sabahirfan:
[CMC-545](https://tools.hmcts.net/jira/browse/CMC-545)

### Change description ###
Add about to start callback for all the events to check whether the event is allowed
If the event is allowed user can progress the event journey
If the selected event is not allowed then user will see the error message on the screen and event journey will not be allowed

- [ ] Yes
- [x] No


dharmendrak (dismissed):

dharmendrak (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/AcknowledgeServiceCallbackHandler.java
```diff
@@ -53,6 +54,10 @@
         return EVENTS;
     }
 
+    public CallbackResponse aboutToStart(CallbackParams callbackParams) {
```
pd-dudkiewicz: Should we call it `emptyCallbackResponse` and put it as protected method in `CallbackHandler`, so it's not copy-pasted in a few places?

sabahirfan: yes, that make sense. done.

pd-dudkiewicz (dismissed):

pd-dudkiewicz (dismissed):

douglasrice (approved):

dharmendrak (approved):

PR #242, opened 2020-10-01
feat: restrict event for ongoing business process on case (CMC-799)

sabahirfan:
### JIRA link (if applicable) ###
[CMC-799](url)

### Change description ###
This PR restricts event if there is ongoing business process on the case. 

**Does this PR introduce a breaking change?** (check one with "x")

- [ ] Yes
- [x] No

PR #240, opened 2020-09-30
feat: generic service notification (CMC-636)

pd-dudkiewicz:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/CMC-639

### Change description ###

Added generic notification to inform solicitors about response to case.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #245, opened 2020-10-01
feat: Propagate CCD events to camunda via polling event emitter (CMC-792)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/CMC-793

### Change description ###

Added polling event emitter:
- searching for all cases with ready business process status
- sending message to camunda triggering specific task for each case
- updating cases business process status to dispatched (if not progressed already)

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/RespondExtensionCallbackHandler.java
```diff
@@ -102,11 +102,12 @@ private CallbackResponse updateResponseDeadline(CallbackParams callbackParams) {
             data.put(RESPONSE_DEADLINE, newDeadline.atTime(MID_NIGHT));
         }
 
-        data.put("businessProcess",
-                 BusinessProcess.builder().activityId("ExtensionResponseHandling").status(READY).build());
+        List<String> errors = new ArrayList<>();
+        businessProcessService.updateBusinessProcess(data, "ExtensionResponseHandling", errors);
```
swalker125: hmm I don't like this updating errors, method is doing quite a lot. I think better if it returned a list of errors or an empty list

src/test/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/CaseEventTaskHandlerTest.java
```diff
@@ -41,10 +40,8 @@ void init() {
         when(mockExternalTask.getActivityId()).thenReturn("activityId");
         when(mockExternalTask.getAllVariables())
             .thenReturn(Map.of(
-                "CCD_ID",
-                CASE_ID.toString(),
-                "CASE_EVENT",
-                NOTIFY_RESPONDENT_SOLICITOR1_FOR_CLAIM_ISSUE.name()
+                "CCD_ID", CASE_ID,
```
swalker125, line 43: think sonar will complain here, one param on each line

swalker125 (commented):
Looks pretty good to me. Only thing I am not a fan of is the updateBusinessProcess method. Think it'll be nicer if it just returns a list of errors though. Let me know thoughts

src/test/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/CaseEventTaskHandlerTest.java
swalker125: oh it's map, ignore 🤦

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/RespondExtensionCallbackHandler.java
pd-dudkiewicz: Didn't want the method to return value and have side effect at the same time. But happy to change it.

swalker125: hmm you're right, it shouldn't return something and change something else. Trying to think how we can split this out

swalker125: should we maybe map businessProcess before sending to service?

Then one method to getProcessErrors(BusinessProcess) which returns errors or empty list. 
Second method which updates businessProccess with ID and ready state?

Then data.put() is still in the handler

Do you think the separate service is necessary or can it exist on the object?

Then we have some fluent code like businessProcess.getErrors(), businessProcess.updateToReadyStatus(activityId)

pd-dudkiewicz: decided to go for returning list of errors - it's the best option imho, as we don't copy-paste the same logic across all handlers

src/main/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/PollingEventEmitterHandler.java
```diff
@@ -0,0 +1,54 @@
...
+        final String taskName = externalTask.getTopicName();
+        log.info("Job {} started", taskName);
+
+        List<CaseDetails> cases = caseSearchService.getCases();
+        log.info("Job '{}' found {} case(s)", taskName, cases.size());
+        cases.stream().map(caseDetailsConverter::toCaseData).forEach(caseData -> {
+            var caseId = caseData.getCcdCaseReference();
+            var businessProcess = caseData.getBusinessProcess();
+            try {
+                runtimeService.createMessageCorrelation("Message" + businessProcess.getActivityId())
```
ckalista: See my comment above, we may want to pass the CCD ID here as Message Name.

src/main/java/uk/gov/hmcts/reform/unspec/handler/DispatchBusinessProcessEventHandler.java
```diff
@@ -0,0 +1,35 @@
...
+@RequiredArgsConstructor
+public class DispatchBusinessProcessEventHandler {
+
+    private final CoreCaseDataService coreCaseDataService;
+
+    @EventListener
+    public void dispatchBusinessProcess(DispatchBusinessProcessEvent event) {
+        BusinessProcess businessProcess = event.getBusinessProcess();
+        if (businessProcess.getStatus() == READY) {
+            coreCaseDataService.triggerEvent(
```
ckalista: I think we have a race condition here we need to be mindful of - can we discuss on Monday?

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -0,0 +1,38 @@
...
+@RequiredArgsConstructor
+public class BusinessProcessService {
+
+    private static final List<BusinessProcessStatus> ERROR_STATUSES = List.of(READY, DISPATCHED, STARTED);
+    private static final String ERROR_MESSAGE = "Business Process Error";
+
+    private final ObjectMapper mapper;
+
+    public List<String> updateBusinessProcess(Map<String, Object> data, String businessProcessEvent) {
+        BusinessProcess businessProcess = mapper.convertValue(data.get("businessProcess"), BusinessProcess.class);
```
ckalista: I have seen we are doing this sort of bespoke conversion in several places. Should we consider using the CaseDetailsConverter instead?

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -0,0 +1,38 @@
...
+        if (Optional.ofNullable(businessProcess)
+            .map(BusinessProcess::getStatus)
+            .map(ERROR_STATUSES::contains)
+            .orElse(false)) {
+            return List.of(ERROR_MESSAGE);
+        }
+        data.put("businessProcess", BusinessProcess.builder().activityId(businessProcessEvent).status(READY).build());
```
ckalista: We need to explicitly reset the `processInstanceId` to null here as well as resetting the `activityId` to `DEFAULT`.

ckalista: I don't think we should be using the activityId as message name to start the business process. Message Name and Activity Id are two distinct concepts in Camunda we should aim to keep it that way. We may have to introduce a new field in `BusinessProcessEvent` to capture the Message Name (CCD EVENT ID)

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -0,0 +1,38 @@
...
+public class BusinessProcessService {
+
+    private static final List<BusinessProcessStatus> ERROR_STATUSES = List.of(READY, DISPATCHED, STARTED);
+    private static final String ERROR_MESSAGE = "Business Process Error";
+
+    private final ObjectMapper mapper;
+
+    public List<String> updateBusinessProcess(Map<String, Object> data, String businessProcessEvent) {
+        BusinessProcess businessProcess = mapper.convertValue(data.get("businessProcess"), BusinessProcess.class);
+        if (Optional.ofNullable(businessProcess)
```
ckalista: I would actually be a even more defensive. Could we check that the business process `STATUS` is either null or `FINISHED` to be permitting a reset to `READY`. The logic is the same as here:
https://github.com/hmcts/unspec-service/pull/242/files#diff-4c2ee2ed8f256169482f5b867f57ea2dR53

Maybe extracting this in a utility class to be shared?

src/main/java/uk/gov/hmcts/reform/unspec/service/search/CaseReadyBusinessProcessSearchService.java
```diff
@@ -0,0 +1,26 @@
...
+@Service
+public class CaseReadyBusinessProcessSearchService extends ElasticSearchService {
+
+    public CaseReadyBusinessProcessSearchService(CoreCaseDataService coreCaseDataService) {
+        super(coreCaseDataService);
+    }
+
+    public Query query(int startIndex) {
+        return new Query(
+            boolQuery().must(matchQuery("data.businessProcess.status", "READY")),
```
ckalista, line 21: Do we need to consider pagination here?

ckalista (changes requested):
Good job!! There are two potential issues I would like to discuss with the whole team. Will set something up after standup on Monday.

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
pd-dudkiewicz: Not sure if there's a point of mapping the whole CaseData if we only need one field value

src/main/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/PollingEventEmitterHandler.java
```diff
@@ -0,0 +1,57 @@
...
+        List<CaseDetails> cases = caseSearchService.getCases();
+        log.info("Job '{}' found {} case(s)", taskName, cases.size());
+        cases.stream().map(caseDetailsConverter::toCaseData).forEach(caseData -> {
+            var caseId = caseData.getCcdCaseReference();
+            var businessProcess = caseData.getBusinessProcess();
+            try {
+                var messageCorrelationBuilder = runtimeService
+                    .createMessageCorrelation(businessProcess.getCamundaEvent()).setVariable("CCD_ID", caseId);
+                Optional.ofNullable(caseData.getStateFlowState())
+                    .ifPresent(flag -> messageCorrelationBuilder.setVariable("STATE_FLOW_STATE", flag));
```
ckalista: Shall we just call it `FLOW_STATE`? Not particular about it though...

ckalista: I don't think we need to pass the `FLOW_STATE` here at all... we can put it in context in each Camunda Service Task that follows instead. Will chat about this with you...

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/RespondToDefenceCallbackHandler.java
```diff
@@ -50,15 +53,15 @@
 
     private CallbackResponse handleNotifications(CallbackParams callbackParams) {
         Map<String, Object> data = callbackParams.getRequest().getCaseDetails().getData();
-        YesOrNo proceeding = mapper.convertValue(data.get(APPLICANT_1_PROCEEDING), YesOrNo.class);
-        var response = mapper.convertValue(data.get("respondent1ClaimResponseType"), DefendantResponseType.class);
-        if (response == FULL_DEFENCE && proceeding == YES) {
-            data.put("businessProcess",
-                     BusinessProcess.builder().activityId("CaseTransferredToLocalCourtHandling").status(READY).build());
+        CaseData caseData = mapper.convertValue(data, CaseData.class);
```
sabahirfan: Should we use CaseDetailsConverter here?

src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
```diff
@@ -88,4 +88,5 @@
     private final CloseClaim discontinueClaim;
 
     private final BusinessProcess businessProcess;
+    private final String stateFlowState;
```
sabahirfan: I don't think we need to preserve flow state, it will be calculated based on the CaseData.

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    private CallbackResponse dispatchBusinessProcess(CallbackParams callbackParams) {
+        Map<String, Object> data = callbackParams.getRequest().getCaseDetails().getData();
+        BusinessProcess businessProcess = mapper.convertValue(data.get("businessProcess"), BusinessProcess.class);
+        if (businessProcess.getStatus() == READY) {
+            data.put("businessProcess", BusinessProcess.builder()
+                .camundaEvent(businessProcess.getCamundaEvent())
+                .status(DISPATCHED)
+                .build());
```
ckalista, line 48: Is this also going to be resetting the other Business Process attributes? Would it be safer to explicitly set the action is and process instance id to null as well here?

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
ckalista: No worries we have discussed this and have since raised a tech debt ticket.

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -0,0 +1,48 @@
...
+                .map(State::getName)
+                .ifPresent(stateName -> data.put("stateFlowState", stateName));
+            return List.of();
+        }
+        return List.of(ERROR_MESSAGE);
+    }
+
+    //TODO: extract once merged:
+    //      https://github.com/hmcts/unspec-service/pull/242/files#diff-4c2ee2ed8f256169482f5b867f57ea2dR53
+    private boolean hasNoOngoingBusinessProcess(BusinessProcess businessProcess) {
```
sabahirfan: This is merged to master, can be extracted now.

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -0,0 +1,48 @@
...
+        return updateBusinessProcess(data, caseEvent, null);
+    }
+
+    public List<String> updateBusinessProcess(Map<String, Object> data, CaseEvent caseEvent, State stateFlowState) {
+        BusinessProcess businessProcess = mapper.convertValue(data.get("businessProcess"), BusinessProcess.class);
+        if (hasNoOngoingBusinessProcess(businessProcess)) {
+            data.put("businessProcess", BusinessProcess.builder().camundaEvent(caseEvent.name()).status(READY).build());
+            Optional.ofNullable(stateFlowState)
+                .map(State::getName)
+                .ifPresent(stateName -> data.put("stateFlowState", stateName));
```
ckalista: Maybe just a suggestion to call this `Flow State`.

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -0,0 +1,48 @@
...
+    private final ObjectMapper mapper;
+
+    public List<String> updateBusinessProcess(Map<String, Object> data, CaseEvent caseEvent) {
+        return updateBusinessProcess(data, caseEvent, null);
+    }
+
+    public List<String> updateBusinessProcess(Map<String, Object> data, CaseEvent caseEvent, State stateFlowState) {
+        BusinessProcess businessProcess = mapper.convertValue(data.get("businessProcess"), BusinessProcess.class);
+        if (hasNoOngoingBusinessProcess(businessProcess)) {
+            data.put("businessProcess", BusinessProcess.builder().camundaEvent(caseEvent.name()).status(READY).build());
```
ckalista, line 26: Should be more explicit, and also reset the other Business Process attributes: activity id & process instance id?

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
ckalista: This is actually quite important. The flow state should never be stored in the case data but always computed on the fly. This is to make it easier once we will be live, to change the state flow engine rules without have to also patch the data.

src/main/java/uk/gov/hmcts/reform/unspec/stateflow/model/State.java
```diff
@@ -22,6 +26,10 @@ public String getName() {
         return name;
     }
 
+    public boolean isFlowState(FlowState flowState) {
```
sabahirfan, line 29: I would try to keep separation between `stateflow` - ( for core state flow that can be reused by others) and `flowstate` - ( where we will have our project specific things). is it possible to have it a utility method in `flowstate`?

ckalista (commented):
Thanks for actioning all my previous suggestions 👍 

A couple of things I would like your view point on. Will chat to you about it.

src/test/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.unspec.callback.CaseEvent.CREATE_CLAIM;
+import static uk.gov.hmcts.reform.unspec.model.BusinessProcessStatus.FINISHED;
+import static uk.gov.hmcts.reform.unspec.model.BusinessProcessStatus.READY;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {BusinessProcessService.class, JacksonAutoConfiguration.class})
+public class BusinessProcessServiceTest {
```
sabahirfan: we can remove `public` from here

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/RespondToDefenceCallbackHandler.java
pd-dudkiewicz: done

src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
pd-dudkiewicz: this is removed now

src/main/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/PollingEventEmitterHandler.java
pd-dudkiewicz: removed

pd-dudkiewicz: removed

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandler.java
pd-dudkiewicz: New object is created here, so anything not set will be null.

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
pd-dudkiewicz: Will do, but leaving it for now until [CMC-636](https://github.com/hmcts/unspec-service/pull/240) gets merged to master, as this branch is based on it.

src/main/java/uk/gov/hmcts/reform/unspec/stateflow/model/State.java
pd-dudkiewicz: Removed the method for now, it's only used in single callback handler. Will add it to FlowState if it's needed in more than one place.

src/test/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessServiceTest.java
pd-dudkiewicz: done

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
pd-dudkiewicz: removed as discussed

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
pd-dudkiewicz: As mentioned above - new object is created here, so activityId and processInstanceId will both be null. But can still set them to null explicitly if we think this is more suitable.

pd-dudkiewicz: @ckalista I've put additional assertions for verifying if they are set to null

ckalista: Good stuff! Thanks 👍

ckalista (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
pd-dudkiewicz: this is now done

src/main/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/PollingEventEmitterHandler.java
```diff
@@ -0,0 +1,54 @@
...
+                    caseId,
+                    businessProcess
+                ));
+            } catch (Exception ex) {
+                log.error(String.format("%s task failed for case: %d, message: %s", taskName, caseId, ex.getMessage()));
+            }
+        });
+
+        externalTaskService.complete(externalTask);
+        log.info("Job '{}' finished", taskName);
```
dharmendrak, line 52: there is an aspect for logging

dharmendrak (dismissed):
LGTM

sabahirfan (dismissed):

src/main/resources/respond_extension.bpmn
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swalker125: do we have tests for these bpmn diagrams?

src/main/java/uk/gov/hmcts/reform/unspec/service/tasks/handler/PollingEventEmitterHandler.java
pd-dudkiewicz: updated

src/main/resources/respond_extension.bpmn
pd-dudkiewicz: pulled them from 793

dharmendrak (dismissed):

sabahirfan (dismissed):

douglasrice (approved):

dharmendrak (approved):

PR #259, opened 2020-10-08
feat: testing support endpoint for reseting business process (CMC-NA)

pd-dudkiewicz:
### Change description ###

- Added testing support endpoint for reseting business process - required for api and e2e tests to pass
- Added setting business process status to ready in callback handlers
- Enabled new endpoint on local, preview and AAT environments
- Updated NoOngoingBusinessProcessAspect to not return error for submitted callbacks

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
+++ b/src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
@@ -21,6 +21,8 @@ import java.util.List;
 import javax.validation.Valid;
 import javax.validation.constraints.PastOrPresent;
 
+import static uk.gov.hmcts.reform.unspec.enums.BusinessProcessStatus.FINISHED;
+
 @Data
 @Builder(toBuilder = true)
 @HasServiceDateTheSameAsOrAfterIssueDate(groups = ConfirmServiceDateGroup.class)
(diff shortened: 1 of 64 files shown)


dharmendrak (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
```diff
@@ -88,4 +90,10 @@
     private final CloseClaim discontinueClaim;
 
     private final BusinessProcess businessProcess;
+
+    public boolean hasNoOngoingBusinessProcess() {
```
sabahirfan, line 94: Do we need to put `@JsonIgnore` for this? so when we serialize this using `CaseDetailsConverter` it's not sent to CCD.

pd-dudkiewicz: It's not required, Jackson uses `isPropertyName` to serialize boolean properties. And even if it wants to serialize it, we're still protected by our tests - they'd fail when callbacks would try to set non-existent case field.

swalker125 (commented):
why do we need to add business process READY to all the about to submit callbacks?

pd-dudkiewicz:
> why do we need to add business process READY to all the about to submit callbacks?

@swalker125 we're sending notifications there

swalker125:
> > why do we need to add business process READY to all the about to submit callbacks?
> 
> @swalker125 we're sending notifications there

as a part of reseting business process with test endpoint?

pd-dudkiewicz:
@swalker125 Ignore me, that change shouldn't be there yet 🤦 reverted the commit

swalker125 (dismissed):

PR #260, opened 2020-10-08
fix: allocated track not persisted (CMC-837 )

sabahirfan:
[CMC-837](https://tools.hmcts.net/jira/browse/CMC-837)

### Change description ###
This PR fixes the bug of not persisting the allocatedTrack.

- [ ] Yes
- [x] No


swalker125 (dismissed):
LGTM

e2e/fixtures/createClaim.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
       }
     },
     claimant: {
+      allocatedTrack: 'SMALL_CLAIM',
```
pd-dudkiewicz: We expect allocatedTrack to be set by service callback, so we shouldn't set it via data store request

pd-dudkiewicz (changes requested):

e2e/fixtures/createClaim.js
sabahirfan: As the same is being used by api tests, and we need this for apit test as well.

sabahirfan: done

pd-dudkiewicz (approved):

dharmendrak (approved):

PR #277, opened 2020-10-12
feat: schedule polling event emitter (CMC-793)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/CMC-793

### Change description ###

- scheduler for polling event emitter, disabled on preview
- added missing bpmn tests.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/unspec/bpmn/AcknowledgeServiceTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+class AcknowledgeServiceTest extends BpmnBaseTest {
+
+    public static final String TOPIC_NAME = "processCaseEvent";
+
+    public AcknowledgeServiceTest() {
+        super("camunda/acknowledge_service.bpmn", "ACKNOWLEDGE_SERVICE");
```
swalker125: can we add /camunda in base class?

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/BpmnBaseTest.java
```diff
@@ -83,6 +84,19 @@ static void shutDown() {
         return engine.getExternalTaskService().createExternalTaskQuery().list();
     }
 
+    /**
+     * Retrieves a process definition which has a start message event with the messageName.
+     *
+     * @param messageName the name of the message.
+     * @return process definitions with given message start message event.
+     */
+    public ProcessDefinition getProcessDefinitionByMessage(String messageName) {
```
swalker125: worth enumerating messageName?

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/CaseStayedSchedulerTest.java
```diff
@@ -38,10 +38,7 @@ void caseStayedSchedulerBmpnShouldFireCaseStayedExternalTask_whenStarted() {
 
         //get external tasks
         List<ExternalTask> externalTasks = getExternalTasks();
-
-        //assert task is as expected
         assertThat(externalTasks).hasSize(1);
-        assertThat(externalTasks.get(0).getTopicName()).isEqualTo("CASE_STAYED_FINDER");
```
swalker125: intentional delete?

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/ClaimantResponseTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    void caseEventTaskShouldFireCaseEventExternalTask_whenStarted() {
+        //assert process has started
+        assertFalse(processInstance.isEnded());
+
+        //assert topic names
+        assertThat(getTopics()).containsOnly(TOPIC_NAME);
+
+        //assert message start event
+        assertThat(getProcessDefinitionByMessage("CLAIMANT_RESPONSE")).isNotNull();
```
swalker125: what does isNotNull tell us here? Can we assert that correct actvityId is started or something?

swalker125: Few other places this is done

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/AcknowledgeServiceTest.java
pd-dudkiewicz: done

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/BpmnBaseTest.java
pd-dudkiewicz: We could use CaseEvents here, which most likely won't change, but I'm not sure if we want to have that dependency. I'd leave it as string, the worst thing that can happen is that definition won't be found.

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/CaseStayedSchedulerTest.java
pd-dudkiewicz: yes, we already have assertions for topic name:
`assertThat(getTopics()).containsOnly(TOPIC_NAME);`
so this line felt redundant

src/test/java/uk/gov/hmcts/reform/unspec/bpmn/ClaimantResponseTest.java
pd-dudkiewicz: It was telling us that there is a process definition triggered by given message. But added checking if process id is correct.

dharmendrak (dismissed):

swalker125 (commented):
Just before I review, is there a reason we haven't gone down the path of one email handler, when camunda supplies the email template reference and we build the data from the reference?

pd-dudkiewicz:
> Just before I review, is there a reason we haven't gone down the path of one email handler, when camunda supplies the email template reference and we build the data from the reference?

@swalker125 I think the idea was that we have one generic external task handler for all ccd events triggered by camunda, where we pass only ccd id and case event id. Also, sending template reference itself wouldn't help too much, because we want to trigger different ccd events for different notifications (in case we'd like to show that it was triggered on the UI).

douglasrice (approved):

dharmendrak (approved):

PR #283, opened 2020-10-13
fix: renamed claim value field and added to pounds method (CMC-899)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-899

### Change description ###

CCD sends back GBP field as pennies. This was throwing off the allocated track logic.
Renamed CCD field and added to pounds method. 

Refactored logic touched by above fix

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/unspec/model/ClaimValue.java
+++ b/src/main/java/uk/gov/hmcts/reform/unspec/model/ClaimValue.java
@@ -12,16 +12,18 @@ import java.math.BigDecimal;
 @Builder
 public class ClaimValue {
 
-    private final BigDecimal statementOfValue;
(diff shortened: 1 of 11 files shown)


pd-dudkiewicz (approved):

src/main/java/uk/gov/hmcts/reform/unspec/model/ClaimValue.java
```diff
@@ -12,16 +12,18 @@
 @Builder
 public class ClaimValue {
 
-    private final BigDecimal statementOfValue;
+    private final BigDecimal statementOfValueInPennies;
```
sabahirfan, line 15: Should we use `BigInteger` for `statementOfValueInPennies` instead?

douglasrice (approved):

PR #307, opened 2020-10-20
refact: CaseData model to be used in callback response (CMC-849)

sabahirfan:
[CMC-849](url)

### Change description ###

The callback handler uses the `CaseData` model to be returned back via its `CallbackResponse`

This PR also removes the duplicate validation logic from the `BusinessProcessService`. As the same is already being invoked in `NoOngoingBusinessProcessAspect` 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/unspec/service/BusinessProcessService.java
```diff
@@ -1,31 +1,26 @@
...
 @RequiredArgsConstructor
 public class BusinessProcessService {
 
-    private static final String ERROR_MESSAGE = "Business Process Error";
-
-    private final ObjectMapper mapper;
-
-    public List<String> updateBusinessProcess(Map<String, Object> data, CaseEvent caseEvent) {
-        CaseData caseData = mapper.convertValue(data, CaseData.class);
+    public CaseData updateBusinessProcess(CaseData caseData, CaseEvent caseEvent) {
```
pd-dudkiewicz, line 15: We still need to pass data map here, otherwise business process will not be set to READY

sabahirfan: good catch. done

src/main/java/uk/gov/hmcts/reform/unspec/aspect/EventEmitterAspect.java
```diff
@@ -20,15 +19,14 @@
...
     private final EventEmitterService eventEmitterService;
 
     @Around("execution(* *(*)) && @annotation(EventEmitter) && args(callbackParams))")
     public Object emitBusinessProcessEvent(ProceedingJoinPoint joinPoint, CallbackParams callbackParams)
         throws Throwable {
         if (callbackParams.getType() == SUBMITTED) {
-            CaseData caseData = caseDetailsConverter.toCaseData(callbackParams.getRequest().getCaseDetails());
-            if (caseData.getBusinessProcess().getStatus() == READY) {
+            CaseData caseData = callbackParams.getCaseData();
+            if (caseData.getBusinessProcess() != null && caseData.getBusinessProcess().getStatus() == READY) {
```
swalker125, line 29: would prefer this logic contained in BusinessProcess but not really a part of your PR so fine to leave I guess!

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/AcknowledgeServiceCallbackHandler.java
```diff
@@ -66,17 +68,17 @@ private CallbackResponse validateDateOfBirth(CallbackParams callbackParams) {
...
 
-        data.put(RESPONSE_DEADLINE, newResponseDate.atTime(MID_NIGHT));
-        List<String> errors = businessProcessService.updateBusinessProcess(data, ACKNOWLEDGE_SERVICE);
+        CaseData caseDataUpdated = businessProcessService.updateBusinessProcess(
+            caseDataBuilder.build(), ACKNOWLEDGE_SERVICE);
 
         return AboutToStartOrSubmitCallbackResponse.builder()
-            .data(data)
-            .errors(errors)
+            .data(caseDetailsConverter.toMap(caseDataUpdated))
```
swalker125, line 81: what about having a toMap method on CaseData?

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/AcknowledgeServiceCallbackHandler.java
```diff
@@ -66,17 +68,17 @@ private CallbackResponse validateDateOfBirth(CallbackParams callbackParams) {
...
-
+        CaseData caseData = callbackParams.getCaseData();
+        LocalDateTime responseDeadline = caseData.getRespondentSolicitor1ResponseDeadline();
         LocalDate newResponseDate = workingDayIndicator.getNextWorkingDay(responseDeadline.plusDays(14).toLocalDate());
+        CaseData.CaseDataBuilder caseDataBuilder = caseData.toBuilder()
+            .respondentSolicitor1ResponseDeadline(newResponseDate.atTime(MID_NIGHT));
 
-        data.put(RESPONSE_DEADLINE, newResponseDate.atTime(MID_NIGHT));
-        List<String> errors = businessProcessService.updateBusinessProcess(data, ACKNOWLEDGE_SERVICE);
+        CaseData caseDataUpdated = businessProcessService.updateBusinessProcess(
```
swalker125, line 77: updateBusinessProcess returning CaseData seems a lot? I would have thought this would be nicer as a void method on BusinessProcess? I'm not actually sure of a use for BusinessProcessService. Again maybe not this PRs concern

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/ConfirmServiceCallbackHandler.java
```diff
@@ -78,11 +80,8 @@
     private CallbackResponse prepopulateServedDocuments(CallbackParams callbackParams) {
         List<ServedDocuments> servedDocuments = List.of(ServedDocuments.CLAIM_FORM);
 
-        Map<String, Object> data = callbackParams.getRequest().getCaseDetails().getData();
-        data.put("servedDocuments", servedDocuments);
-
         return AboutToStartOrSubmitCallbackResponse.builder()
-            .data(data)
+            .data(Map.of("servedDocuments", servedDocuments))
```
swalker125: is the point of this PR not to remove bits like this?

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/ConfirmServiceCallbackHandler.java
```diff
@@ -119,34 +118,29 @@ private CallbackResponse prepareCertificateOfService(CallbackParams callbackPara
...
-        CaseData caseDateUpdated = caseData.toBuilder()
+        CaseData.CaseDataBuilder caseDataBuilder = caseData.toBuilder()
             .deemedServiceDateToRespondentSolicitor1(deemedDateOfService)
-            .respondentSolicitor1ResponseDeadline(responseDeadline)
-            .build();
+            .respondentSolicitor1ResponseDeadline(responseDeadline);
 
         CaseDocument certificateOfService = certificateOfServiceGenerator.generate(
-            caseDateUpdated,
+            caseDataBuilder.build(),
```
swalker125, line 132: really don't like the requirement to build case data here, is hard to know what is needed for generate method. Again - maybe not your PRs concern 😅

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandler.java
```diff
@@ -36,17 +40,19 @@
...
         if (businessProcess.getStatus() == READY) {
-            data.put("businessProcess", BusinessProcess.builder()
-                .camundaEvent(businessProcess.getCamundaEvent())
-                .status(DISPATCHED)
-                .build());
+            caseDataBuilder
+                .businessProcess(BusinessProcess.builder()
+                                     .camundaEvent(businessProcess.getCamundaEvent())
+                                     .status(DISPATCHED)
+                                     .build());
```
swalker125, line 51: good example of why businessProcessService is now not needed

swalker125: also another use case for some void methods to update business process in case data directly rather than need case data builder

src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
```diff
@@ -30,7 +31,9 @@
 @HasServiceDateTheSameAsOrAfterIssueDate(groups = ConfirmServiceDateGroup.class)
 public class CaseData {
 
+    @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
     private final Long ccdCaseReference;
+    @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
     private final CaseState ccdState;
```
swalker125, line 37: why are these added to CaseData?

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandlerTest.java
```diff
@@ -32,23 +36,33 @@
...
+                .extracting("camundaEvent")
+                .isEqualTo("testCamundaEvent");
+            assertThat(response.getData())
+                .extracting("businessProcess")
+                .extracting("activityId")
+                .isNull();
+            assertThat(response.getData())
+                .extracting("businessProcess")
+                .extracting("processInstanceId")
+                .isNull();
```
swalker125: does .extracting not take multiple strings for assertions / can you not chain from .extracting business process

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/RequestExtensionCallbackHandlerTest.java
```diff
@@ -155,7 +141,7 @@ void shouldUpdateResponseDeadlineToProposedDeadline_whenExtensionAlreadyAgreed()
             var response = (AboutToStartOrSubmitCallbackResponse) handler.handle(params);
 
             assertThat(response.getData())
-                .containsEntry(RESPONSE_DEADLINE, proposedDeadline.atTime(MID_NIGHT));
+                .containsEntry(RESPONSE_DEADLINE, proposedDeadline.atTime(MID_NIGHT).toString());
```
swalker125, line 144: suspicious toString here

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/RequestExtensionCallbackHandlerTest.java
```diff
@@ -178,9 +164,18 @@ void shouldNotUpdateResponseDeadline_whenExtensionIsNotAlreadyAgreed() {
...
-            verify(businessProcessService).updateBusinessProcess(new HashMap<>(), REQUEST_EXTENSION);
+            assertThat(response.getData())
+                .extracting("businessProcess")
+                .extracting("camundaEvent")
+                .isEqualTo(REQUEST_EXTENSION.name());
+
+            assertThat(response.getData())
+                .extracting("businessProcess")
+                .extracting("status")
+                .isEqualTo("READY");
```
swalker125: again I think you can extract multiple / chain here

swalker125 (commented):
Couple of things, maybe not a part of your PR though.

- I don't see point of BusinessProcessService and it's needed in a few places
- I think case data should have a toMap method rather than requiring CaseDetailsConvertor everywhere. Whether this is a method in CaseData or inherited from another class I don't know. What do you reckon? We could have a base type like DocmosisData that has a toMap method
- Why aren't we just using setters directly on CaseData, is it cleaner than using toBuilder everywhere?

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/AcknowledgeServiceCallbackHandler.java
sabahirfan: I agree that `BusinessProcessService` seems unnecessary and we can clean that up as separate PR not part of this.

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/AcknowledgeServiceCallbackHandler.java
sabahirfan: I will prefer the domain objects plain java objects, don't want to inject `ObjectMapper` bean there.

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/ConfirmServiceCallbackHandler.java
sabahirfan: will visit this later if needed.

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/RequestExtensionCallbackHandlerTest.java
sabahirfan: assertion fails as in `response.getData()` map it's type is String and we have `LocalDateTime` object, so have to make both of same type.

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandlerTest.java
sabahirfan: I am afraid not. as we have condition on one property. we can not chain after that.

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/ConfirmServiceCallbackHandler.java
sabahirfan: This is mid callback where we perpopulate the one field for the particular page. And this is the only use case for this kind of functionality. Sending the complete caseData doesn't seem to be right in this scenario.

src/main/java/uk/gov/hmcts/reform/unspec/model/CaseData.java
sabahirfan: These fields are being used already and we need them in our business logic. But we don't want to save these as part of data map in CaseDetails as we don't have any case field for these.

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/RequestExtensionCallbackHandlerTest.java
sabahirfan: I am afraid not. as we have condition on one property. we can not chain after that.

src/main/java/uk/gov/hmcts/reform/unspec/aspect/EventEmitterAspect.java
sabahirfan: added null safety to the existing condition.

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandlerTest.java
sabahirfan: done

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/RequestExtensionCallbackHandlerTest.java
sabahirfan: done

dharmendrak (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/ConfirmServiceCallbackHandler.java
sabahirfan: to be consistent across all changed that as well.

dharmendrak (approved):

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/AcknowledgeServiceCallbackHandler.java
swalker125: hmm we could have this in an interface as we have for DocmosisData however?

swalker125 (approved):

PR #318, opened 2020-10-27
fix:resolve case data model mismatch(CMC-NA)

dharmendrak:


douglasrice (approved):

swalker125 (dismissed):
why changes to unspecified docker

sabahirfan (approved):
I don't think we need submodule changes to be merged. otherwise LGTM.

PR #341, opened 2020-11-02
feat: move case to stayed after claimant response (CMC-645)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-645

### Change description ###

Update content in submitted callback
Move case to stayed after RESPOND TO DEFENCE event

```
[ ] Yes
[x] No
```


pd-dudkiewicz (dismissed):

dharmendrak (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/service/flowstate/StateFlowEngine.java
```diff
@@ -70,6 +70,7 @@ public StateFlow build() {
                 .transitionTo(CLAIM_DISCONTINUED).onlyIf(claimDiscontinued)
                 .transitionTo(CLAIM_WITHDRAWN).onlyIf(claimWithdrawn)
             .state(FULL_DEFENCE)
+                .transitionTo(CLAIM_STAYED).onlyIf(claimantRespondToDefence)
```
sabahirfan: Do we need this predicate `.onlyIf(claimantRespondToDefence)` ? we can transition to without this, otherwise I would expect the predicate to be unique there.
```suggestion
                .transitionTo(CLAIM_STAYED)
```

sabahirfan:
By looking at the story, I have a question.

- The case will be moved to **stayed** in scenario when claimant solicitor will choose intention of **not to proceed**. 

- The case will be moved to **stayed** in scenario when claimant solicitor will choose intention of ** to proceed**.  **Is this correct?**

swalker125:
> By looking at the story, I have a question.
> 
>     * The case will be moved to **stayed** in scenario when claimant solicitor will choose intention of **not to proceed**.
> 
>     * The case will be moved to **stayed** in scenario when claimant solicitor will choose intention of ** to proceed**.  **Is this correct?**

From what Luke was saying earlier, yes. The case will move offline for release one and into stayed state in CCD

src/main/java/uk/gov/hmcts/reform/unspec/service/flowstate/StateFlowEngine.java
```diff
@@ -70,6 +70,7 @@ public StateFlow build() {
                 .transitionTo(CLAIM_DISCONTINUED).onlyIf(claimDiscontinued)
                 .transitionTo(CLAIM_WITHDRAWN).onlyIf(claimWithdrawn)
             .state(FULL_DEFENCE)
+                .transitionTo(CLAIM_STAYED)
```
sabahirfan, line 73: Do we want to have different internal states based on the path it took to reach CCD stayed state? 

- Scheduler moves the case to stayed
- claimant solicitor does not want to proceed, and we move this to stayed.

ckalista: That's a really good question @sabahirfan - I will review the flow states this week hopefully, also we have a story in the backlog to refactor all the states. We will pick this up again, for now whatever is implemented here is ok.

ckalista (approved):

src/main/java/uk/gov/hmcts/reform/unspec/service/flowstate/FlowPredicate.java
```diff
@@ -37,7 +37,7 @@
             && caseData.getApplicant1DefenceResponseDocument() != null;
 
     public static final Predicate<CaseData> schedulerStayClaim = caseData ->
-        caseData.getCcdState() == STAYED;
+        caseData.getCcdState() == STAYED && caseData.getDeemedServiceDateToRespondentSolicitor1() == null;
```
sabahirfan, line 40: should we use the caseData filed that we populate in full defence rather than confirm service?

dharmendrak (approved):

PR #349, opened 2020-11-04
fix: dispatch business process only on event emitter (CMC-944)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-944

### Change description ###

About to start callback on dispatch business process to not trigger if business process is started

Boolean flag on emit business process event set to true from polling event emitter, and false from submitted callbacks of events.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandler.java
@@ -12,14 +12,17 @@ import uk.gov.hmcts.reform.unspec.helpers.CaseDetailsConverter;
 import uk.gov.hmcts.reform.unspec.model.BusinessProcess;
 import uk.gov.hmcts.reform.unspec.model.CaseData;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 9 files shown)


dharmendrak (dismissed):

pd-dudkiewicz (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/DispatchBusinessProcessCallbackHandler.java
```diff
@@ -31,14 +34,30 @@
...
     @Override
     public List<CaseEvent> handledEvents() {
         return EVENTS;
     }
 
+    private CallbackResponse checkIfBusinessProcessStarted(CallbackParams callbackParams) {
+        CaseData caseData = callbackParams.getCaseData();
+        List<String> errors = new ArrayList<>();
+
+        if (caseData.getBusinessProcess().getStatusOrDefault() == STARTED) {
```
ckalista: Actually here I would be less strict with the check. What it Camunda already fully processed the workflow event, i.e. Status is then `FINISHED` already... 

I would say better be more conservative, check that the status IS NOT `READY` instead?

ckalista (dismissed):
Hiya, just one point to consider, otherwise spot on 👍

ckalista (approved):

douglasrice (approved):

PR #377, opened 2020-11-13
feat: code changes for camunda claim form (CMC-934)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-934

### Change description ###

Code changes for camunda claim form generation

```
[ ] Yes
[ ] No
```


ccd-definition/CaseEvent.json
```diff
@@ -389,12 +389,26 @@
...
     "CallBackURLAboutToSubmitEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/cases/callbacks/about-to-submit",
     "ShowSummary": "N",
     "ShowEventNotes": "N"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "ID": "GENERATE_CLAIM_FORM",
+    "Name": "Generate claim form",
+    "Description": "Generates claim form after successful payment",
```
ckalista, line 406: I have run past our suggestion with Emily already - she agreed with our approach. The event description we want to see in the history is going to be something along the lines of `Claim issued` (which comes after `Claim created`).  `CASE_ISSUED` is also going to be renamed to something like `AWAITING_CONFIRM_SERVICE`.

All of which will be part of another story.

ckalista (dismissed):

ccd-definition/CaseEvent.json
swalker125: just for our knowledge, name is added on the event history, description I belive is what would be read out on a screen reader (and when you hover over the event in next step dropdown)

PR #417, opened 2020-11-26
build: bumped spring boot to 2.4.0 (CMC-1052)

pd-dudkiewicz:

```
[ ] Yes
[x] No
```


sabahirfan:
With springboot 2.4.0 we don't need to explicitly exclude `junit-vintage-engine` from `spring-boot-starter-test` Please cleanup that as well.

sendgrid-client/build.gradle
```diff
@@ -57,6 +57,5 @@ dependencies {
   testImplementation libraries.junit5
   testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test', {
     exclude group: 'junit', module: 'junit'
```
sabahirfan, line 59: I think this is also excluded already, please verify.

dharmendrak (dismissed):

sendgrid-client/build.gradle
pd-dudkiewicz: yes, removed

swalker125 (approved):

sabahirfan (approved):

PR #419, opened 2020-11-27
refactor: separate case events json files in definition (cmc-1038)

sabahirfan:
[CMC-1038](https://tools.hmcts.net/jira/browse/CMC-1038)

### Change description ###
This PR splits one CaseEvent json file into related separate Json files under CaseEvent directory.

- [ ] Yes
- [x] No


ccd-definition/CaseEvent/Camunda/BusinessProcessEvents.json
```diff
@@ -0,0 +1,47 @@
+[
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "ID": "DISPATCH_BUSINESS_PROCESS",
+    "Name": "Dispatch business process",
+    "Description": "Set READY business process status to DISPATCHED",
+    "DisplayOrder": 22,
```
pd-dudkiewicz: Should we remove `DisplayOrder` for any event that's not user event?

swalker125: Agree on the above, a big benefit of this is that we can accurately keep track of where events will be displayed for the user. A further improvement could be to split user events by state, but not a part of this :)

dharmendrak (dismissed):

ccd-definition/CaseEvent/Camunda/BusinessProcessEvents.json
sabahirfan: Sounds fine as it is irrelevant. It's not part of this tech debt. but I can take care of that as well.

swalker125: @sabahirfan I would ignore state for now, as there is a big state refactor coming :)

pd-dudkiewicz (approved):

douglasrice (approved):

PR #430, opened 2020-12-02
feat: pba payment screen (CMC-1062)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-1062

### Change description ###

PRD client
Dynamic list for PBA accounts
Store fee data in CCD

Removal of old fixed list PBA field will break all exisiting cases in all envs

```
[x] Yes
[] No
```


ccd-definition/CaseField.json
```diff
@@ -118,6 +118,57 @@
...
+    "FieldType": "Label",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "ID": "applicantSolicitor1PbaAccountsIsEmpty",
+    "Label": "Hide pba and show noPbaAccountLabel",
+    "FieldType": "YesOrNo",
+    "SecurityClassification": "Public",
```
pd-dudkiewicz, line 159: There was this conversation a while ago about having `_Category` on case fields. I don't think we agreed on approach for all new fields, but this one could be marked as a `WORKAROUND`.

e2e/pages/createClaim/paymentReference.page.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = {
+
+  fields: {
+    paymentReference: '#paymentReference',
+  },
+
+  async updatePaymentReference() {
+    I.waitForElement(this.fields.paymentReference);
+    I.clearField(this.fields.paymentReference);
```
pd-dudkiewicz, line 11: why is this required?

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandler.java
```diff
@@ -74,6 +87,34 @@ private CallbackResponse validateDateOfBirth(CallbackParams callbackParams) {
...
+        Optional<SolicitorReferences> references = ofNullable(caseData.getSolicitorReferences());
+        String paymentReference = ofNullable(caseData.getPaymentReference())
+            .orElse(references.map(SolicitorReferences::getApplicantSolicitor1Reference).orElse(""));
+
+        String authToken = callbackParams.getParams().get(BEARER_TOKEN).toString();
+        List<String> pbaNumbers = getPbaAccounts(authToken);
+
+        CaseData.CaseDataBuilder caseDataBuilder = caseData.toBuilder()
+            .claimFee(feesService.getFeeDataByClaimValue(caseData.getClaimValue()))
+            .applicantSolicitor1PbaAccounts(DynamicList.fromList(pbaNumbers))
```
pd-dudkiewicz, line 110: We should probably check if there's some value selected, otherwise it will be removed when going to previous page and visiting this one again. Might be useful thing to have in `DynamicList` class.

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandler.java
```diff
@@ -74,6 +87,34 @@ private CallbackResponse validateDateOfBirth(CallbackParams callbackParams) {
...
+            .data(caseDetailsConverter.toMap(caseDataBuilder.build()))
+            .build();
+    }
+
+    private List<String> getPbaAccounts(String authToken) {
+        Optional<Organisation> organisation = organisationService.findOrganisation(authToken);
+
+        List<String> pbaNumbers = new ArrayList<>();
+        organisation.ifPresent(org -> pbaNumbers.addAll(org.getPaymentAccount()));
+        return pbaNumbers;
```
pd-dudkiewicz: How about:
```
return organisationService.findOrganisation(authToken)
    .map(Organisation::getPaymentAccount)
    .orElse(List.of());
```

src/main/java/uk/gov/hmcts/reform/unspec/service/FeesService.java
```diff
@@ -40,15 +34,13 @@ private FeeLookupResponseDto lookupFee(ClaimValue claimValue) {
...
 
-    private BigInteger getFeeAmountInPence(FeeLookupResponseDto feeLookupResponseDto) {
-        var feeAmountPounds = feeLookupResponseDto.getFeeAmount();
-
-        return feeAmountPounds.multiply(PENCE_PER_POUND).toBigInteger();
-    }
+    private Fee buildFeeDto(FeeLookupResponseDto feeLookupResponseDto) {
+        BigDecimal calculatedAmount = feeLookupResponseDto.getFeeAmount()
+            .multiply(PENCE_PER_POUND)
+            .setScale(0, RoundingMode.CEILING);
```
pd-dudkiewicz: I think we should always use `RoundingMode.UNNECESSARY`. When there's going to be some weird value it's better to throw exception than charge more (even if it's only one pence).

src/main/java/uk/gov/hmcts/reform/unspec/service/OrganisationService.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final OrganisationApi organisationApi;
+    private final AuthTokenGenerator authTokenGenerator;
+
+    public Optional<Organisation> findOrganisation(String authToken) {
+        try {
+            return ofNullable(organisationApi.findUserOrganisation(authToken, authTokenGenerator.generate()));
+
+        } catch (FeignException.NotFound | FeignException.Forbidden ex) {
+            log.error("User not registered in MO", ex);
```
pd-dudkiewicz, line 28: Is there some potentially useful data in response body? If yes, then it's better to log `ex.contentUTF8()`, because FeignException does not show it by default.

ccd-definition/CaseField.json
swalker125: I'll be honest I wasn't sure about that from the get go, I think if naming and comments can explain a field sufficiently then that should do. However if @AlisonRevittHM asks for it then I will add.

e2e/pages/createClaim/paymentReference.page.js
swalker125: To be honest, it isn't. The field is prepopulated from solicitor references, the clear was so that we type in new text

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandler.java
swalker125: much nicer

src/main/java/uk/gov/hmcts/reform/unspec/service/FeesService.java
swalker125: sorry I think you mentioned this too me before, will update.

src/main/java/uk/gov/hmcts/reform/unspec/service/OrganisationService.java
swalker125: Possible errors from this endpoint are:

400 | An invalid ID was provided
403 | Forbidden Error: Access denied
404 | No Organisation found with the given ID

We know that 400 in theory should never happen from a logged in user, 403 in theory should also never happen when we are whitelisted but the user could maybe not have the right roles to query this endpoint, so this just leaves the 404. I don't think we are swollowing anything useful there so will leave as is.

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandler.java
swalker125: I'm not sure how to get this working, the selected value is gone when I click back, maybe one to look at in tech debt?

pd-dudkiewicz: if it's not simple to fix then yeah, we could get that in separate ticket

src/test/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandlerTest.java
```diff
@@ -141,6 +161,80 @@ void shouldReturnNoError_whenSoleTraderDateOfBirthIsInThePast() {
...
+                    feeData.getCode(),
+                    feeData.getVersion()
+                );
+
+            assertThat(getDynamicList(response))
+                .isEqualTo(DynamicList.builder().value(DynamicListElement.EMPTY).build());
+        }
+
+        private DynamicList getDynamicList(AboutToStartOrSubmitCallbackResponse response) {
+            ObjectMapper mapper = new ObjectMapper();
```
pd-dudkiewicz: this could be moved to field, so we don't create new instance on every method call

pd-dudkiewicz (dismissed):

src/main/java/uk/gov/hmcts/reform/unspec/handler/callback/CreateClaimCallbackHandler.java
swalker125: https://tools.hmcts.net/jira/browse/CMC-1122

ccd-definition/CaseField.json
swalker125: This can always be retrofitted as well, has no functional effect for the story

pd-dudkiewicz (dismissed):

pd-dudkiewicz (dismissed):

pd-dudkiewicz (dismissed):

pd-dudkiewicz (approved):

douglasrice (approved):

PR #449, opened 2020-12-11
feat: idam stub support (CMC-894)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/CMC-894

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseState/systemUpdate.json
```diff
@@ -20,13 +20,6 @@
     "UserRole": "caseworker-civil-systemupdate",
     "CRUD": "CRUD"
   },
-  {
```
swalker125: why is this removed?

pd-dudkiewicz: This is to allow `Solicitor responds to defence` e2e test to pass locally while using idam stub. In idam stub we have one user with all roles, so when case moved to `STAYED` state, it was still available to user, instead of redirecting to Case List. We could change last test assertion if the stub is used, but I couldn't find a reason why we need this permission.

sabahirfan (dismissed):

ccd-definition/AuthorisationCaseState/systemUpdate.json
swalker125: I don't think it's safe to remove, for example for my 202 story. CLAIMANT_RESPONSE event leaves case in stayed state from CCD definition, we then need to process events such as notifications etc. System update user needs to see case to trigger events

swalker125: surely there are already notification events that need to happen already too?

sabahirfan (approved):

PR #484, opened 2021-01-05
fix: assert dynamic list list items are as expected (CMC-1111)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-1111

### Change description ###

- New global midEventFieldForPage array with id and dynamic list boolean.
- assertValidData checks the array to see if it contains the current pageId being asserted against.
- If the pageId has a mid event then we need to add those fields to case data regardless.
- If pageId links to dynamic list field in array then remove UUID (code) and assert the response data labels are equal to expected.

```
[ ] Yes
[x] No
```


e2e/api/steps.js
```diff
@@ -188,7 +196,10 @@ const assertValidData = async (data, pageId) => {
   const responseBody = await response.json();
 
   assert.equal(response.status, 200);
-  addIssueClaimMidEventFields(pageId, responseBody);
+
+  if (pagesWithMidEvents.includes(pageId)) {
```
pd-dudkiewicz: `midEventFieldForPage.hasOwnProperty(pageId)` could be used here, so we don't need `pagesWithMidEvents`

pd-dudkiewicz (dismissed):

pd-dudkiewicz (approved):

douglasrice (approved):

PR #500, opened 2021-01-08
fix: added min value validation to claim value (CMC-1155)

pd-dudkiewicz:
### Change description ###

The service returned an error when claim value was set to 0£, as there's no fee available.

```
[ ] Yes
[x] No
```


swalker125 (approved):

swalker125 (changes requested):
could you update this to 100? Should solve the other bug https://tools.hmcts.net/jira/browse/CMC-1157

Validated with Darren that a £1 minmum is ok

swalker125 (approved):

dharmendrak (approved):

sabahirfan (approved):

douglasrice (approved):

PR #512, opened 2021-01-13
feat: enable share a case (CMC-1124)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/CMC-1124

```
[ ] Yes
[x] No
```


ccd-definition/CaseField.json
```diff
@@ -752,5 +752,13 @@
     "Label": "Defendant response date",
     "FieldType": "Date",
     "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "UNSPECIFIED_CLAIMS",
+    "ID": "localAuthorityPolicy",
```
swalker125: guess local authority doesn't make sense for us. These will need to be named defendant / claimant + 1 / 2 + organisationPolicy

src/main/java/uk/gov/hmcts/reform/unspec/enums/CaseRole.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.hmcts.reform.unspec.enums;
+
+import lombok.Getter;
+
+@Getter
+public enum CaseRole {
+    SOLICITOR;
```
swalker125: These case roles will need to be DEFENDANT / CLAIMANT + SOLICITOR + 1 / 2

src/main/java/uk/gov/hmcts/reform/unspec/model/OrganisationPolicy.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class OrganisationPolicy {
+
+    @JsonProperty("Organisation")
+    private Organisation organisation;
```
swalker125: Could we name this type OrganisationId or something to stop the clash with Organisation type when used?

src/main/java/uk/gov/hmcts/reform/unspec/model/Organisation.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.Data;
+import lombok.extern.jackson.Jacksonized;
+
+@Data
+@Builder
+@Jacksonized
+public class Organisation {
```
swalker125: Odd they decided to create this as an object, seems unnecessary 🤷‍♂️

ccd-definition/CaseField.json
pd-dudkiewicz: And where do we populate these 4 fields? It's not quite clear to me how these fields are going to be used 🤔

swalker125: I think these will all exist in the create a claim event. Unsure how design will implement a "this case has 2 defendants" etc, but in my mind all the fields will optionally show based on some information entered earlier. With the 4 separate fields we can hard code the case roles for each org policy, so that when I invite an org through the UI as a user, any solicitors assigned via the caa will be given the correct role

src/main/java/uk/gov/hmcts/reform/unspec/model/Organisation.java
pd-dudkiewicz: more details coming in the future? 🤔

ccd-definition/CaseField.json
pd-dudkiewicz: How do we get all organisations ids in create claim event if we search it using single auth token?

swalker125: The org policy field is actually the org search used for respondent journey. When the event is started xui makes a call to prd fetching ALL orgs. The user will search and pick one, adding the id.

swalker125: I guess FPLs use case is slightly different, as I don't think they are implementing the respondent journey stuff quite yet. As they are onboarding local authorities they will want to add the id of the org the user belongs to probably in a callback at case creation stage right? We will not need to do this, because when mixed with respondent journey, the user will be entering this information for us.

pd-dudkiewicz: that makes much more sense now, thanks 👍 updated the PR

src/main/java/uk/gov/hmcts/reform/unspec/model/OrganisationId.java
```diff
@@ -0,0 +1,15 @@
...
+import lombok.Data;
+import lombok.extern.jackson.Jacksonized;
+
+@Data
+@Builder
+@Jacksonized
+public class OrganisationId {
+
+    @JsonProperty("OrganisationID")
+    private String organisationID;
```
sabahirfan: the only field and class name are same, do we really need this class?

pd-dudkiewicz: This is how the field is structured in CCD.

sabahirfan: May be we can rename the class property then and leaving the @JsonProperty as it is.

sabahirfan (dismissed):

sabahirfan (dismissed):

swalker125 (dismissed):

swalker125 (approved):

sabahirfan (approved):

PR #549, opened 2021-01-28
fix: start business process diagram

sabahirfan:
N/A

### Change description ###
Abort was not working with start business process in case of BpmnError in handler. Have to remove the internal abort from the diagram so the error bubbles up to user diagram.

- [ ] Yes
- [x] No


swalker125:
hmm, I think we had this tested in bpmn unit tests. If this isn't working we may need to update the tests too. Thoughts?

swalker125 (approved):

PR #560, opened 2021-02-04
tech: remove duplicate flows for same internal state (CMC-1287)

sabahirfan:
[CMC-1287](https://tools.hmcts.net/jira/browse/CMC-1287)

### Change description ###
This PR  removes duplicate flows for same internal state `PROCEEDS_WITH_OFFLINE_JOURNEY`.

- [ ] Yes
- [x] No


swalker125:
@sabahirfan do we actually care about two different proceeds with offline journey states? I assume once in this state it will never come into the live service again, so we will never use this data. Thoughts?

@ckalista

sabahirfan:
Yes we care in internal states/flow states, with CCD states we don't care. 

Each internal state name should define only one possible flow, while CCD state can have more than one paths to reach there. 

i.e. we have to populate the event names from case data object, we are going to use the state flow engine to tell us what path it took, based on that we have to prepare sequence of events for RPA. This is just one use case there will be more.

swalker125 (approved):

dharmendrak (approved):

pd-dudkiewicz (approved):

PR #566, opened 2021-02-05
feat: capture claimant email and idam id (CMC-1201)

swalker125:

https://tools.hmcts.net/jira/browse/CMC-1201

### Change description ###

Capture claimant email and idam id.
Query idam, display email
User can select if they want to proceed with this email or enter another

```
[ ] Yes
[x] No
```


ccd-definition/ComplexTypes/IdamUserDetails.json
```diff
@@ -0,0 +1,16 @@
+[
```
ckalista: I wonder if we should model this slightly differently... 
In my mind the technical data we pull from IDAM (email & id) should be stored in an IDAM specific complex type e.g., `applicantSolicitor1IdamUserDetails` while the business data we care about from a business point of view should be stored in a complex type that doesn't contain any technical data specific to IDAM: `applicantSolicitor1EmailAddress` .The `applicantSolicitor1IdamUserDetails` data  should only be kept on the case for a relatively short amount of time: (1) clear `applicantSolicitor1IdamUserDetails.email` as soon as we no longer need it, i.e. when it has been copied across into `applicantSolicitor1EmailAddress.email`. (2) clear `applicantSolicitor1IdamUserDetails.id` in Camunda immediately after case roles have been fixed up (@dharmendrak story would take care of that).

ckalista (commented):
Hiya Sam, I have commented on a possible suggestion for improvement...

dharmendrak (dismissed):

swalker125:
Screenshot_2021-02-15 Create a case - HM Courts Tribunals Service - GOV UK(2)
Screenshot_2021-02-15 Create a case - HM Courts Tribunals Service - GOV UK(1)

douglasrice (approved):

dharmendrak (approved):

PR #568, opened 2021-02-08
fix: bumped dependencies for CVE-2020-29582 and added suppression

swalker125:

n/a

### Change description ###

bumped dependencies:

-  dependency group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-common', version: '1.4.30'
-  dependency group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-jdk7', version: '1.4.30'
-  dependency group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-jdk8', version: '1.4.30'
-  dependency group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib', version: '1.4.30'
-  dependency group: 'org.jetbrains.kotlin', name: 'kotlin-reflect', version: '1.4.30'

Added suppression for false positive (https://github.com/jeremylong/DependencyCheck/issues/2785)

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,6 +75,12 @@ allprojects {
       dependencySet(group: 'com.google.guava', version: '30.1-jre') {
         entry 'guava'
       }
+      dependency group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-common', version: '1.4.30'
+      dependency group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-jdk7', version: '1.4.30'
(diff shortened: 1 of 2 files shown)


douglasrice (approved):

dharmendrak (approved):

sabahirfan (approved):
do we need to bump dependencies? as we are ignoring the CVE anyway.

swalker125:
> do we need to bump dependencies? as we are ignoring the CVE anyway.

bumping dependencies to latest to fix issue (issue is fixed in 1.4.21 and above) but also need to suppress due to false positive. Better to not have the security issue and suppress rather than have the issue and suppress I think!

PR #594, opened 2021-02-17
fix: updated e2e test locators for collection document

pd-dudkiewicz:
### Change description ###

Upload document element sitting inside collection does not have index anymore.

```
[ ] Yes
[x] No
```


swalker125 (approved):
Do we need to run a specific version of definition store / data store for this change?

pd-dudkiewicz:
@swalker125 It's not related to definition or data stores, it's only a change in xui. But it will require pulling latest xui image so the tests will pass locall.

sabahirfan (approved):

PR #595, opened 2021-02-17
refactor: change court selection from court location to court code  (CMC-1177)

devmarcy:
https://tools.hmcts.net/jira/projects/CMC/issues/CMC-1177

### Change description ###
Changed court location to Court code
Added a link to court finder for claimant
Added a link for defendant response which directs to the claimant-chosen court

```
[ ] Yes
[x] No
```


ccd-definition/ComplexTypes/CourtLocation.json
```diff
@@ -3,15 +3,17 @@
     "ID": "CourtLocation",
     "ListElementCode": "applicantPreferredCourt",
     "FieldType": "Text",
```
swalker125, line 5: thoughts on this being FieldType: Number?

swalker125 (changes requested):

ccd-definition/ComplexTypes/CourtLocation.json
AlisonRevittHM: Is text with the regex better and more future proofed than number?

devmarcy: It is currently text with regex @AlisonRevittHM

devmarcy: oh, it seems I wrote Text to Number by mistake on the commit message - I meant from Number to Text!

AlisonRevittHM: no worries. glad we're of the same viewpoint. cheers

ccd-definition/ComplexTypes/CourtLocation.json
```diff
@@ -3,15 +3,16 @@
...
-    "ElementLabel": "Court name",
+    "ElementLabel": "Court Code",
     "HintText": "This should be your preferred County Court hearing centre.",
-    "SecurityClassification": "Public"
+    "SecurityClassification": "Public",
+    "RegularExpression": "[0-9]{3}"
```
swalker125, line 9: could you update api tests just to check this works?

swalker125 (changes requested):
one more thing, really good stuff though!

ccd-definition/ComplexTypes/CourtLocation.json
ckalista: So what are we saying here in the end.. Should it be changed to Number?

ckalista (changes requested):

e2e/api/steps.js
```diff
@@ -254,13 +255,14 @@ const assertValidData = async (data, pageId) => {
...
-const assertCallbackError = async (pageId, eventData, expectedErrorMessage) => {
+const assertCallbackError = async (pageId, eventData, expectedErrorMessage, responseBodyMessage = 'Unable to proceed because there are one or more callback Errors or Warnings' ) => {
   const response = await apiRequest.validatePage(eventName, pageId, {...caseData, ...eventData}, 422);
   const responseBody = await response.json();
-
   assert.equal(response.status, 422);
-  assert.equal(responseBody.message, 'Unable to proceed because there are one or more callback Errors or Warnings');
-  assert.equal(responseBody.callbackErrors[0], expectedErrorMessage);
+  assert.equal(responseBody.message, responseBodyMessage);
+  if(responseBody.callbackErrors != null){
```
pd-dudkiewicz, line 254: is it null when we pass value that does not match regex?

devmarcy: @ckalista I think it is much better for it to be Text+regex, also when using Number, field interpolation doesn't work

@pd-dudkiewicz This was done so that it could assert the error correctly, it is null because instead of giving a callback error its just returning "Case data validation failed" - so to answer your question even if we pass a value that does match the regex assuming there's no other callback errors it should also be null

ccd-definition/ComplexTypes/CourtLocation.json
swalker125: @ckalista there were implications using interpolation with a number field. So text with regex

e2e/api/steps.js
swalker125: callback errors is only populated for callbacks, aka validation in our code. This regex validation is not in the callback and therefor this block is empty. Works with existing tests, we can change if the need requires

e2e/api/steps.js
```diff
@@ -52,6 +52,8 @@ module.exports = {
 
     await assertCorrectEventsAreAvailableToUser(config.solicitorUser, 'AWAITING_CASE_NOTIFICATION');
     await assertCorrectEventsAreAvailableToUser(config.defendantSolicitorUser, 'AWAITING_CASE_NOTIFICATION');
+    await assertCallbackError('Court', data[eventName].invalid.Court.courtLocation.applicantPreferredCourt,
```
swalker125: this method name is now incorrect, we are not asserting a callback error. Better way to do this could be an entirely new method that just calls assertCallbackError inside? Or maybe renaming existing method, thoughts?

e2e/fixtures/events/createClaim.js
```diff
@@ -141,9 +141,17 @@ module.exports = {
       PaymentReference: {
         paymentReference: 'Applicant reference'
       }
+    },
+    invalid:{
+      Court: {
+        courtLocation: {
+          applicantPreferredCourt: '3a3'
```
swalker125: should we also check if longer than 3 characters? Can the court code be shorter? aka 21, or would it be 021?

swalker125 (commented):
Couple more bits about testing :)

e2e/api/steps.js
devmarcy: Could we rename the method to just assertError?

e2e/fixtures/events/createClaim.js
devmarcy: following the regex pattern being used ( "[0-9]{3}") it would have to be 021 in that case

e2e/fixtures/events/createClaim.js
```diff
@@ -156,9 +156,17 @@ module.exports = {
...
+        courtLocation: {
+          applicantPreferredCourt: ['3a3','21','3333']
+        }
+      }
     }
   },
   createClaimLitigantInPerson: {
     valid: createClaimData('No')
   },
+
```
swalker125: can remove this line

swalker125 (dismissed):
Comment about a blank line added, but otherwise LGTM

e2e/api/steps.js
pd-dudkiewicz: I'd say it's better to have a separate method for that to have clear separation between callback and data store errors. We could even use it later for testing other ccd validation bits that are not covered now, like min/max values.

ckalista (dismissed):

swalker125 (dismissed):

e2e/api/steps.js
```diff
@@ -57,8 +57,15 @@ module.exports = {
 
     await assertCorrectEventsAreAvailableToUser(config.solicitorUser, 'AWAITING_CASE_NOTIFICATION');
     await assertCorrectEventsAreAvailableToUser(config.defendantSolicitorUser, 'AWAITING_CASE_NOTIFICATION');
+    let i;
+    for(i=0; i<data[eventName].invalid.Court.courtLocation.applicantPreferredCourt.length; i++) {
+      await assertError('Court', data[eventName].invalid.Court.courtLocation.applicantPreferredCourt[i],
```
pd-dudkiewicz, line 60: Can we please have `null` as 3rd argument here, instead of `'The data entered is not valid for this type of field.'`?

It's currently a bit misleading because we actually don't assert it in `assertError` method.

pd-dudkiewicz (changes requested):

e2e/api/steps.js
devmarcy: all done!

pd-dudkiewicz (dismissed):

ckalista (dismissed):

ccd-definition/ComplexTypes/DQ/RequestedCourt.json
```diff
@@ -16,12 +23,21 @@
   },
   {
     "ID": "RequestedCourt",
-    "ListElementCode": "name",
-    "FieldType": "Text",
-    "ElementLabel": "Court name",
+    "ListElementCode": "courtFinderLabel",
+    "FieldType": "Label",
+    "ElementLabel": "<a target=\"_blank\" href=\"courttribunalfinder.service.gov.uk/search/address\">Use court finder</a> (opens in a new window) to find the 3 digit court code of your preferred court.",
```
douglasrice: You're missing https:// at the front of the courttribunalfinder url

douglasrice (changes requested):

ccd-definition/ComplexTypes/DQ/RequestedCourt.json
devmarcy: Hi Doug, this has been added now👍🏽

douglasrice (approved):

PR #623, opened 2021-02-25
feat: hearing tab (CMC-1272)

devmarcy:

https://tools.hmcts.net/jira/browse/CMC-1272

### Change description ###

Created 'Hearings' tab for Case View

```
[ ] Yes
[x] No
```


devmarcy:
image

swalker125 (changes requested):
We're going to need to change the headings to differentiate applicant / respondent. Sorry didn't notice this earlier. Also please update screenshot with hearing support field filled :)

AlisonRevittHM:
Is there an order as to which litigant is shown first where there are 2 DQs?
Also, is this only for HMCTS staff?

AlisonRevittHM:
Also, is this future-proofed for multi party?

dharmendrak (dismissed):

devmarcy:
updated screenshot 
image

swalker125:
> Is there an order as to which litigant is shown first where there are 2 DQs?
> Also, is this only for HMCTS staff?

@AlisonRevittHM please save these questions for the refinement session we want to have on these tabs. This is just a first pass to facilitate some conversation. From the prototype I did lots of context was lost when passed to design, so the content attached to tickert 1272 isn't quite up to date any more, it also doesn't match some of our data models.

Currently the tabs are not fixed for any roles, every role will see this.

The way tabs work in ccd defintion is, you add a field to a specified tab. In terms of being future proofed, we just add more fields to the tabs.

swalker125 (changes requested):
you've updated the labels in CaseField but not CaseEventToField, we don't want to change the events in any way for this ticket.

dharmendrak (approved):

PR #658, opened 2021-03-08
fix: added case navigation for RPA nightly tests (CMC-1255)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/CMC-1255

### Change description ###

- goToCase navigation for RPA nightly test - it's required for checking if no events are available once scheduler moves case offline (the page refresh doesn't work in codeceptjs, so need to navigate via UI)
- configurable waitForTimeout value - environment running nightly pipeline is slower than other pipelines - this might improve nightly build flakiness

```
[ ] Yes
[x] No
```


swalker125 (approved):
lgtm

dharmendrak (approved):

PR #684, opened 2021-03-15
tech: add missing unit tests for IsEventAllowedOnFlowState (CMC-1179)

sabahirfan:
[CMC-1179](https://tools.hmcts.net/jira/browse/CMC-1179)

### Change description ###
Add missing unit tests for `IsEventAllowedOnFlowState`.

- [ ] Yes
- [x] No


pd-dudkiewicz (commented):
How about reworking this test to use `@ArgumentsSource`, where we'd provide flow state with list of expected allowed case events? That way we could also assert if any other state outside that collection is not allowed. Plus we could use enum values instead of strings in `@CsvSource`, which makes it easier for refactoring.

dharmendrak (dismissed):

pd-dudkiewicz (approved):

dharmendrak (approved):

